PR #2, opened 2019-08-09
build tries,longest palindrome, longest Word done-pls review

PremalSamale:


LongestWordinDictionary.java
```diff
@@ -0,0 +1,96 @@
...
+	                }
+	                for (Node nei: node.children.values()) {
+	                    stack.push(nei);
+	                }
+	            }
+	        }
+	        return ans;
+	    }
+	}
+}
```
ashmeetdeol19: You have made the code more complex then it actually is.
Think about easy implementation of Tries.

PR #10, opened 2019-09-27
Please review.

Jaysparkexel:
Trie was not taught yesterday. All three problems of trie would be taught today in the class.


Jaysparkexel:
@sehaj-grover done problem 2.

PR #31, opened 2020-01-03
Update trie.java

sakshi1705:


trie.java
```diff
@@ -0,0 +1,77 @@
+search complexity is o(n), o(1) for each character
+the root and its childfen and its nested children are stored in a hashmap and searching in a hashmap for a single character is o(1)
+space : o(1)
+class TrieNode {
```
neha79091: Space complexity is  nk. where n is number of words in list to be inserted in trie and k is length of longest word in dictionart

PR #33, opened 2020-01-05
Implemented Trie Problems

dasgprat:
Please review the PR and also let me know in case of any issue or concerns.

### DIFF
--- a/Problem2.java
+++ b/Problem2.java
@@ -1,0 +1,88 @@
+/**
+LeetCode Submitted : YES
+Time Complexity : O(Num of words * Max Height of Tree)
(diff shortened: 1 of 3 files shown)


Problem2.java
```diff
@@ -0,0 +1,88 @@
+/**
+LeetCode Submitted : YES
+Time Complexity : O(Num of words * Max Height of Tree)
+Space Complexity : O(Num of words)
+
```
neha79091: Space complexity will be O(n*k) where n is number of words in list to be inserted and k is length of longest word in dictionary.

PR #36, opened 2020-01-11
tries-1

samhithaaaa:


longestword.py
```diff
@@ -0,0 +1,51 @@
+# Runtime - O(m*n) where m is # of words and n is length of longest word
+    # Memory - O(# of words)
+
```
neha79091: Space complexity will be O(n*k) where n is number of words in list to be inserted and k is length of longest word in dictionary.

PR #41, opened 2020-02-03
Solutions for tries - Please review

poojakittur:
Time and space complexities written at the beginning of files. Each of them executed successfully on leetcode


poojakittur:
Since each node of trie can have only 26 children, wont it be considered as constant space?

neha79091:
> Since each node of trie can have only 26 children, wont it be considered as constant space?

Yes, you are right

PR #52, opened 2020-02-14
Completed Sprint - Please Review

gayathrig269:


gayathrig269:
Please review

gayathrig269:
Please Review

Gayathri_Ganesh_Tries_1.cpp
```diff
@@ -0,0 +1,82 @@
...
+            }
+            this->isEnd=false;
+        } //empty constructor
+    }; //end of class TrieNode
+public:
+    /** Initialize your data structure here. */
+    Trie() {}
+    
+    TrieNode *root=NULL; //initialize root to NULL
+    /** Inserts a word into the trie. */
```
super30admin: root has to be initialized as` new TrieNode. `

Gayathri_Ganesh_Tries_1.cpp
```diff
@@ -0,0 +1,82 @@
...
+        public:
+        bool isEnd;
+        TrieNode *children[26]; //26 as only lowercase letters
+        TrieNode() {        //constructor to initilaize children[] and isEnd members
+            for(int i=0;i<26;i++)
+            {
+                this->children[i]=NULL;
+            }
+            this->isEnd=false;
+        } //empty constructor
```
super30admin: This part is redundant we don't need it. Like in class we have an empty constructor in trienode

super30admin (commented):
Search, insert and starts with seem fine. I am posting the correct code for your reference
`class Trie {
public:
    struct TrieNode {
        TrieNode* nodes[26] = {nullptr};
        bool isEnd = false;
    };
    
    TrieNode* root = nullptr;
    
    /** Initialize your data structure here. */
    Trie() {
        root = new TrieNode();
    }
    
    /** Inserts a word into the trie. */
    void insert(string word) {
        TrieNode* root_temp = root;
        for(int i = 0; i < word.length(); i++){
            if(root_temp->nodes[word[i] - 'a'] == nullptr){
                root_temp->nodes[word[i] - 'a'] = new TrieNode();
            }
            root_temp = root_temp->nodes[word[i] - 'a'];
        }
        root_temp->isEnd = true;
    }
    
    /** Returns if the word is in the trie. */
    bool search(string word) {
        TrieNode* root_temp = root;
        for(int i = 0; i < word.length(); i++){
            if(root_temp->nodes[word[i] - 'a'] == nullptr){
                return false;
            }
            else{
                root_temp = root_temp->nodes[word[i] - 'a'];
            }
        }
        return root_temp->isEnd;
    }
    
    /** Returns if there is any word in the trie that starts with the given prefix. */
    bool startsWith(string prefix) {
        TrieNode* root_temp = root;
        for(int i = 0; i < prefix.length(); i++){
            if(root_temp->nodes[prefix[i] - 'a'] == nullptr){
                return false;
            }
            else{
                root_temp = root_temp->nodes[prefix[i] - 'a'];
            }
        }
        return true;
    }
};

/**
 * Your Trie object will be instantiated and called as such:
 * Trie* obj = new Trie();
 * obj->insert(word);
 * bool param_2 = obj->search(word);
 * bool param_3 = obj->startsWith(prefix);
 */`

super30admin:
Kindly review the in line comments also

gayathrig269:
Completed Tries-1. Please review.

PR #66, opened 2020-04-23
Adding code for 3 questions

RutujaKaushike:


devashishnyati:
Nice try Rutuja!
Try longest word in dictionary and replace words with Trie.

PR #71, opened 2020-04-29
problem-1 completed

schakrad:


problem-1.java
```diff
@@ -0,0 +1,85 @@
+//TC:O(N*K) where k is the length of longest word
```
devashishnyati, line 1: Can you explain the time complexity?

devashishnyati:
Good job Sindhu!
Submit other problems as well.

PR #83, opened 2020-06-05
Tries-1

Siyami:


p3.js
```diff
@@ -0,0 +1,78 @@
...
+    'r': 17,
+    's': 18,
+    't': 19,
+    'u': 20,
+    'v': 21,
+    'w': 22,
+    'x': 23,
+    'y': 24,
+    'z': 25
+}
```
super30admin, line 43: We need this hashmap. We can simply take the asciii code in js as well

super30admin:
Good Job Siyami

PR #95, opened 2020-07-02
COmpleted Trie-1. Kindly review and share feedback.

Sumesh-H:


Implement Trie.java
```diff
@@ -0,0 +1,65 @@
+//Time Complexity O(n)
+//Space Complexity O(1) since number of alphabets is constant in Trie. i.e. each TrieNode can have only 26 children
```
AroopN, line 2: Even though the number of alphabets is constant, how many trie nodes do we need ?

AroopN:
Good work, Sumesh.

AroopN:
Good work, Sumesh.

Implement Trie.java
Sumesh-H: It will depend on the input string, correct?

AroopN: yes

> It will depend on the input string, correct?

Yes, the space complexity of storing the strings in Trie can be O(L) where L is the length of all the strings. But search/ startswith can be done in O(1) space after trie has been implemented.

Sumesh-H: Got it, thanks.

PR #96, opened 2020-07-03
Completed Tries-1 : Please review.

shantanu609:


AbhiniveshP:
Nice work on this PR Shantanu!!!
What if you're asked to extend lower case characters to upper case characters and some special symbols. In that case, it'll be difficult for you to modify this approach of using an array as indexing the characters might turn out to be an issue. So, also try implementing Trie using a HashMap in a TrieNode instead of array of size 26.

AbhiniveshP:
Nice work on this PR Shantanu!!!
What if you're asked to extend lower case characters to upper case characters and some special symbols. In that case, it'll be difficult for you to modify this approach of using an array as indexing the characters might turn out to be an issue. So, also try implementing Trie using a HashMap in a TrieNode instead of array of size 26.

PR #109, opened 2020-07-08
Completed all the problems, please review.

samarthjj:


AbhiniveshP:
Nice work on all these Trie questions Samarth!!!
What if you are asked that you're allowed more characters instead of lower case letters and you have to extend your solution involving Trie Node to these other characters as well? How would you extend your solution? Is usage of an array of size 26 good in that case? How would indexing work if more characters are involved?

AbhiniveshP:
Nice work on all these Trie questions Samarth!!!
What if you are asked that you're allowed more characters instead of lower case letters and you have to extend your solution involving Trie Node to these other characters as well? How would you extend your solution? Is usage of an array of size 26 good in that case? How would indexing work if more characters are involved?

PR #167, opened 2020-10-23
Completed Tries1 - Please review

Sneha9p:


Tries1.java
```diff
@@ -0,0 +1,43 @@
+//Time complexity - O(M) - M is the key length
+//Space complexity - O(M)
```
samarthjj, line 2: Space, would be O(M* L), where L is is the average length of the words in the Trie.

samarthjj:
Good job Sneha, Please also complete the other 2 problems as well.

samarthjj:
Good job Sneha, Please also complete the other 2 problems as well.

samarthjj:
Good job Sneha, Please review the inline comments.
Please also complete the other 2 problems as well.

samarthjj:
Good job Sneha, Please review the inline comments.
Please also complete the other 2 problems as well.

PR #190, opened 2020-11-24
Tries done

GayatriHungund81194:


trie1.py
```diff
@@ -0,0 +1,43 @@
+#Time Complexity: O(length)
```
samarthjj, line 1: Time complexity to insert, search and starts with is O(L)

trie1.py
```diff
@@ -0,0 +1,43 @@
...
+#Space Complexity: O(length)
```
samarthjj, line 2: Space occupied by the Tries DS is O(n*L), where n is the number words in the tree and L is the average length of all the words.

trie2.py
```diff
@@ -0,0 +1,56 @@
+# As taught in class, using trie to solve the problem
+# Time Complexity: O(n)
+# Space Complexity: O(n)
```
samarthjj, line 3: Time and Space: O(n * L)

samarthjj:
Good work Gayatri, Please review the inline comments.

samarthjj:
Good work Gayatri, Please review the inline comments.

PR #466, opened 2022-05-11
Hi Team,

AayushSanghai:
Implement Trie prefix and Replace Word Done. Please review.

### DIFF
--- a/ImplemenTriePrefixTree.java
+++ b/ImplemenTriePrefixTree.java
@@ -1,0 +1,67 @@
+class Trie {
+
+    //Time Complexity : 0(n) where n:length of the word
+    //Space Complexity: 0(n)
+    //Did it successfully run on leetcode: Yes
+    //Did you face any problem while coding: No
+
+    //In brief explain your approach
+
+    class TrieNode{ //Creation of a Trie DS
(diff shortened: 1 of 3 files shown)


ImplemenTriePrefixTree.java
```diff
@@ -0,0 +1,67 @@
+class Trie {
+
+    //Time Complexity : 0(n) where n:length of the word
+    //Space Complexity: 0(n)
```
shrivatsav, line 4: Please check the space complexity.

shrivatsav (commented):
Good Work!

shrivatsav:
Good Work!

ImplemenTriePrefixTree.java
AayushSanghai: It is correct as per me. Can you please point out the additional space I am using?

shrivatsav: To store one word of length l, it is `O(l)`. You have `n` words.

AayushSanghai: Ok,
So the space complexity will be 0(ln)?

PR #498, opened 2022-06-10
solved problem-Please review

jigar99:


jeelpatel18 (commented):
Good work. Please submit 3rd problem also. Write the TC and SC too.

jeelpatel18:
Good work. Please submit 3rd problem also. Write the TC and SC too.

PR #543, opened 2022-08-02
Completed Tries-1 Please Review

sahithi6445:


designTrie.java
```diff
@@ -0,0 +1,73 @@
+// Time Complexity :O(L) where L is length of word or prefix
+// Space Complexity :O(1)
```
daredevilteja, line 2: for insertion it is O(m) where m is the new nodes inserted

daredevilteja:
Good Work! Keep it up

PR #547, opened 2022-08-02
Tire-1

maharshi71:
I have a little bit of confusion in the time complexities can you please clarify it ? 
For example for the longest word in bfs we will be doing all the nodes so we are going through all the char in the words array right? so do that mean it is going to be sum(Len(word)) ?

### DIFF
--- a/ReplaceWords.py
+++ b/ReplaceWords.py
@@ -1,0 +1,50 @@
+class TrieNode:
+    def __init__(self):
+        self.isEnd = False
+        self.children = [None] * 26
+
+
+class Trie:
+    def __init__(self):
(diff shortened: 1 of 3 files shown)


keshav-Gupta-IS (commented):
For Longest word problem you can say O(n*k) where n is the no. of words and k is the average length of word. The main thing while explaining the complexities is to breakdown and explain why you came up with the expression.

keshav-Gupta-IS:
For Longest word problem you can say O(n*k) where n is the no. of words and k is the average length of word. The main thing while explaining the complexities is to breakdown and explain why you came up with the expression.

PR #551, opened 2022-08-03
Completed Sprint Tries-1

siddbanerjee2012:


implement_trie.java
```diff
@@ -0,0 +1,56 @@
+// Time Complexity :O(mn)
```
jeelpatel18, line 1: TC will be length of string

implement_trie.java
```diff
@@ -0,0 +1,56 @@
...
+// Space Complexity : O(n)
```
jeelpatel18, line 2: n*k where k is avg Length of word

jeelpatel18 (commented):
Well Done.

jeelpatel18:
Well Done.

PR #716, opened 2023-10-11
Kindly Review - Tries-1

apurvashinde357:


Nandha2Tech (commented):
Great Work!
For Insert in trie TC: O(n*l) SC: O(n*l) n is the number of words and l is the average length of each word. For search O(l) length of the search word.
For replace words also TC and SC would be O(n*l)

Nandha2Tech:
Great Work!
For Insert in trie TC: O(n*l) SC: O(n*l) n is the number of words and l is the average length of each word. For search O(l) length of the search word.
For replace words also TC and SC would be O(n*l)

PR #772, opened 2024-03-04
Tries 1

Riyazmd007:


Nandha2Tech (commented):
Great Work! Kindly include TC and SC for all problems. For longest_word_dictionary problem you are sorting on the words array which would add extra TC. Instead try to go over the words in trie alone and find the longest word.

Nandha2Tech:
Great Work! Kindly include TC and SC for all problems. For longest_word_dictionary problem you are sorting on the words array which would add extra TC. Instead try to go over the words in trie alone and find the longest word.

PR #774, opened 2024-03-07
Done with Tries-1

Prinitha:


LongestWordInDictionary.py
```diff
@@ -0,0 +1,46 @@
+'''
+TC: O(n) where n is all the words given to build Trie
```
ishika2741, line 2: What about the length of the word

ishika2741 (commented):
Good work!

ishika2741:
Good work!

PR #847, opened 2024-12-13
Completed Tries - 1

KousalyaK03:


Erlybird:
The first problem, length of the longestword is not optimal solution. Try to solve the question in Trie, that's the objective of this question.

Erlybird:
The first problem, length of the longestword is not optimal solution. Try to solve the question in Trie, that's the objective of this question.

PR #903, opened 2025-03-14
Tries 1 Done!

PranayKalva:


Nandha2Tech (commented):
Great Work! To find the longest word in the dictionary, kindly try to optimize the logic by applying bfs/dfs on the trie to get the longest word. In the implemented logic here, sorting is done on the dictionary which will be inefficient.

Nandha2Tech:
Great Work! To find the longest word in the dictionary, kindly try to optimize the logic by applying bfs/dfs on the trie to get the longest word. In the implemented logic here, sorting is done on the dictionary which will be inefficient.

PR #975, opened 2025-08-06
Completed Tries-1

joelkny97:


super30admin:
Strengths:
- The student has a good understanding of the Trie data structure and its operations.
- The code is clean, well-commented, and follows best practices.
- The student correctly handles memory management with destructors.
- The time and space complexity analyses are accurate.

Areas for Improvement:
- In Exercise_1.cpp, the checks for `curr == nullptr` in each method are unnecessary since the root is always initialized in the constructor. These checks can be removed to simplify the code.
- In Exercise_2.cpp, the BFS implementation could be more clearly explained or commented, as it currently assumes the reader understands the logic behind traversing from the end of the alphabet backwards.
- In Exercise_3.cpp, the `cout` statement in the `getRoot` method should be removed as it is likely a debugging leftover and not needed in the final solution.

PR #978, opened 2025-08-25
slved 3 questions and creating a PR to merge

nsh2207:


super30admin:
Strengths:
- The code is clean, modular, and well-commented.
- The time and space complexity analyses are accurate and clearly stated.

Areas for Improvement:
- In Question_1.cpp, the `get_node` function could be made a private member function of the Trie class for better encapsulation.
- In Question_2.cpp, the `findLargest` function could be optimized further by pruning paths that cannot possibly lead to a longer word than the current best.
- The use of a fixed-size vector (26 elements) for children nodes assumes lowercase English letters. This should be documented or made configurable if other character sets are needed.

PR #984, opened 2025-09-24
added solution for tries-1

satvikapadakanti:


super30admin:
- **Strengths**:
  - The student demonstrates a strong understanding of Trie data structures and their applications.
  - The code is modular, with clear separation of concerns.
  - The implementations are efficient and correct.

- **Areas for Improvement**:
  - In `longest.java`, the `TreeNode` class could be made more generic by using a map for children instead of an array, which would allow handling non-lowercase letters or other characters.
  - In `replace.java`, the `shorter` method could be optimized further by breaking early if no root is found, but the current implementation is already efficient.
  - The student could add comments to explain the purpose of each method and the overall logic, especially for more complex operations like DFS in `longest.java`.

PR #989, opened 2025-10-01
Add Trie-based solutions for dictionary problems

ravithakur2k:
Added three new Python files implementing Trie data structures: LongestWordDictionary.py for finding the longest word with Trie and BFS/DFS, ReplaceWords.py for efficient prefix replacement in sentences, and TriePrefixTree.py for basic Trie operations (insert, search, startsWith). These solutions improve prefix search and manipulation efficiency.

### DIFF
--- a/LongestWordDictionary.py
+++ b/LongestWordDictionary.py
@@ -1,0 +1,69 @@
+#Time complexity: O(N * L) for insert, search and startsWith where n is the number of words and L in the length of each word
+#Space Complexity: O(n) for using the array data structure for children
+# Did this code successfully run on Leetcode : Yes
+# Any problem you faced while coding this : No
+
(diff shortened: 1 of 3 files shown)


super30admin:
- Strengths:
  - Correct implementation of the Trie data structure in `TriePrefixTree.py`.
  - Good use of comments to explain the purpose and complexity of each method.
  - Demonstrates understanding of Trie applications in different problems (`LongestWordDictionary.py` and `ReplaceWords.py`).
- Areas for Improvement:
  - Consistency in implementation details across different files (e.g., using an array vs. a dictionary for children nodes).
  - The `insert` method in `ReplaceWords.py` has an early termination condition (`if curr.isWord: break`), which is not standard for a general Trie implementation and should be documented or justified.
  - The `bfs` method in `LongestWordDictionary.py` could be more clearly explained, especially the reverse iteration (`for i in range(25, -1, -1)`).

PR #1007, opened 2026-01-29
MY_Solution

SrinivasaSupreeth:


super30admin:
It seems there has been a misunderstanding. You were asked to implement a Prefix Tree (Trie) for the problem, but you have submitted solutions for two other problems (Replace Words and Longest Word in Dictionary). These problems might use a Trie in their optimal solutions, but you have not implemented a Trie here.

For the problem "Implement Prefix Tree", you need to create a class called Trie with the following methods:
- __init__(): Initializes the trie.
- insert(word): Inserts a word into the trie.
- search(word): Returns True if the word is in the trie.
- startsWith(prefix): Returns True if there is any word in the trie that starts with the prefix.

I recommend revisiting the problem statement and implementing the Trie class. You can use a dictionary to represent the children of each node, and each node should have a flag to mark the end of a word.

Here is a basic structure to get you started:

class TrieNode:
    def __init__(self):
        self.children = {}
        self.is_end = False

class Trie:
    def __init__(self):
        self.root = TrieNode()

        current = self.root
        for char in word:
            if char not in current.children:
                current.children[char] = TrieNode()
            current = current.children[char]
        current.is_end = True

        current = self.root
        for char in word:
            if char not in current.children:
                return False
            current = current.children[char]
        return current.is_end

    def startsWith(self, prefix: str) -> bool:
        current = self.root
        for char in prefix:
            if char not in current.children:
                return False
            current = current.children[char]
        return True

This implementation has O(L) time complexity for each operation, where L is the length of the word or prefix.

PR #1008, opened 2026-01-30
Completed Implementing Prefix Tree and Replacing words problem

thondaha:
Implemented Prefix Tree in PrefixTree.java class 
Implemented ReplaceWord problem in ReplaceWords.java class


super30admin:
Your implementation of the PrefixTree (Trie) in `PrefixTree.java` is excellent. It correctly implements all required methods with optimal time and space complexity. The code is clean and well-documented. The test cases you provided are helpful for verifying correctness.

One minor suggestion: in the `startsWith` method, you might consider adding a comment that it returns true for the empty string (as your test case shows) which is correct because every word starts with the empty string. This is a good edge case to handle.

Regarding `ReplaceWords.java`: This is a good extension and demonstrates a practical use of the Trie. However, note that the problem only asked for the Trie implementation. If this were an assignment, you should ensure that you are solving the exact problem stated. But since you provided both, it shows your understanding.

In `ReplaceWords.java`, the method `getShorterVersionOfTheWord` is implemented correctly. However, you might want to refactor the loop to check the current node after moving to the child. Alternatively, you can break as soon as you find a root. Your current approach is efficient.

Keep up the good work!

PR #1009, opened 2026-01-30
2problems completed

BharathVuppala96:


super30admin:
Your implementation of the Trie in "Implementation Trie.py" is correct and efficient. You have correctly used a fixed array of size 26 for children, which is suitable for lowercase English letters. The code is clean and easy to understand. Well done!

However, note that the problem only requires the implementation of the Trie class. The "Replace Words.py" file contains code for a different problem and should not be included when submitting the solution for the Trie problem. Make sure to only submit the code that is directly relevant to the problem.

One minor suggestion: You can define the TrieNode class inside the Trie class to make it an inner class, which can help with encapsulation and avoid polluting the global namespace. For example:

class Trie:
    class TrieNode:
        def __init__(self):
            self.children = [None] * 26
            self.is_end = False

    def __init__(self):
        self.root = self.TrieNode()

But this is optional and your current structure is acceptable.

Overall, your Trie implementation is excellent.

